Issue #116, opened 2022-04-22
rocketmq-replicator adapt to the new connect api


Slideee:

Slideee:
@odbozhou could you please review the code :)

odbozhou:
> @odbozhou could you please review the code :)

It looks very good, I will review it soon

PR #117, opened 2022-04-22
[ISSUE #116] replicator adapt to the new connect api

Slideee:

(https://github.com/apache/rocketmq-connect/issues/116)

## Brief changelog

replicator adapt to the new connect api

## Verifying this change

XXXX

### DIFF
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
@@ -19,17 +19,14 @@ package org.apache.rocketmq.replicator;
 
 import com.alibaba.fastjson.JSONObject;
 import io.openmessaging.KeyValue;
-import io.openmessaging.connector.api.data.DataEntryBuilder;
-import io.openmessaging.connector.api.data.EntryType;
+import io.openmessaging.connector.api.component.task.source.SourceTask;
+import io.openmessaging.connector.api.component.task.source.SourceTaskContext;
+import io.openmessaging.connector.api.data.ConnectRecord;
 import io.openmessaging.connector.api.data.Field;
 import io.openmessaging.connector.api.data.FieldType;
 import io.openmessaging.connector.api.data.Schema;
-import io.openmessaging.connector.api.data.SourceDataEntry;
-import io.openmessaging.connector.api.source.SourceTask;
-import java.nio.ByteBuffer;
-import java.nio.charset.StandardCharsets;
+import io.openmessaging.connector.api.data.SchemaBuilder;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -43,6 +40,7 @@ import org.apache.rocketmq.replicator.config.ConfigUtil;
 import org.apache.rocketmq.replicator.config.TaskConfig;
 import org.apache.rocketmq.replicator.offset.OffsetSyncStore;
 import org.apache.rocketmq.replicator.schema.FieldName;
+import org.apache.rocketmq.replicator.schema.SchemaEnum;
 import org.apache.rocketmq.tools.admin.DefaultMQAdminExt;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -64,8 +62,18 @@ public class MetaSourceTask extends SourceTask {
     }
 
     @Override
-    public void start(KeyValue config) {
+    public void validate(KeyValue config) {
+
+    }
+
+    @Override
+    public void init(KeyValue config) {
         ConfigUtil.load(config, this.config);
+    }
+
+    @Override
+    public void start(SourceTaskContext sourceTaskContext) {
+        super.start(sourceTaskContext);
 
         try {
             this.srcMQAdminExt = Utils.startMQAdminTool(this.config);
@@ -78,22 +86,25 @@ public class MetaSourceTask extends SourceTask {
         this.started = true;
     }
 
-    @Override public void stop() {
+    @Override
+    public void stop() {
         if (started) {
             started = false;
         }
         srcMQAdminExt.shutdown();
     }
 
-    @Override public void pause() {
+    @Override
+    public void pause() {
 
     }
 
-    @Override public void resume() {
+    @Override
+    public void resume() {
 
     }
 
-    @Override public Collection<SourceDataEntry> poll() {
+    @Override public List<ConnectRecord> poll() {
         log.debug("polling...");
         List<String> groups = JSONObject.parseArray(this.config.getTaskGroupList(), String.class);
 
@@ -106,7 +117,7 @@ public class MetaSourceTask extends SourceTask {
             }
             return Collections.emptyList();
         }
-        List<SourceDataEntry> res = new ArrayList<>();
+        List<ConnectRecord> res = new ArrayList<>();
         for (String group : groups) {
             ConsumeStats stats;
             try {
@@ -117,38 +128,19 @@ public class MetaSourceTask extends SourceTask {
             }
 
             for (Map.Entry<MessageQueue, OffsetWrapper> offsetTable : stats.getOffsetTable().entrySet()) {
-
                 MessageQueue mq = offsetTable.getKey();
                 long srcOffset = offsetTable.getValue().getConsumerOffset();
                 long targetOffset = this.store.convertTargetOffset(mq, group, srcOffset);
 
+                List<Field> fields = new ArrayList<Field>();
+                Schema schema = new Schema(SchemaEnum.OFFSET.name(), FieldType.STRING, fields);
+                schema.getFields().add(new Field(0, FieldName.OFFSET.getKey(), SchemaBuilder.string().build()));
+
                 JSONObject jsonObject = new JSONObject();
-                jsonObject.put(RmqConstants.NEXT_POSITION, srcOffset);
-
-                Schema schema = new Schema();
-                schema.setDataSource(this.config.getSourceRocketmq());
-                schema.setName(mq.getTopic());
-                schema.setFields(new ArrayList<>());
-                schema.getFields().add(new Field(0,
-                    FieldName.OFFSET.getKey(), FieldType.INT64));
-
-                DataEntryBuilder dataEntryBuilder = new DataEntryBuilder(schema);
-                dataEntryBuilder.timestamp(System.currentTimeMillis())
-                    .queue(this.config.getStoreTopic())
-                    .entryType(EntryType.UPDATE);
-                dataEntryBuilder.putFiled(FieldName.OFFSET.getKey(), targetOffset);
-
-                SourceDataEntry sourceDataEntry = dataEntryBuilder.buildSourceDataEntry(
-                    ByteBuffer.wrap(RmqConstants.getPartition(
-                        mq.getTopic(),
-                        mq.getBrokerName(),
-                        String.valueOf(mq.getQueueId())).getBytes(StandardCharsets.UTF_8)),
-                    ByteBuffer.wrap(jsonObject.toJSONString().getBytes(StandardCharsets.UTF_8))
-                );
-                String targetTopic = new StringBuilder().append(group).append("-").append(mq.getTopic())
-                    .append("-").append(mq.getQueueId()).toString();
-                sourceDataEntry.setQueueName(targetTopic);
-                res.add(sourceDataEntry);
+                jsonObject.put(FieldName.OFFSET.getKey(), targetOffset);
+                ConnectRecord connectRecord = new ConnectRecord(Utils.offsetKey(mq.getTopic(), mq.getBrokerName(), String.valueOf(mq.getQueueId())),
+                    Utils.offsetValue(srcOffset), System.currentTimeMillis(), schema, jsonObject.toJSONString());
+                res.add(connectRecord);
             }
         }
         return res;
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
@@ -133,7 +133,7 @@ public class MetaSourceTask extends SourceTask {
                 long targetOffset = this.store.convertTargetOffset(mq, group, srcOffset);
 
                 List<Field> fields = new ArrayList<Field>();
-                Schema schema = new Schema(SchemaEnum.OFFSET.name(), FieldType.STRING, fields);
+                Schema schema = new Schema(SchemaEnum.OFFSET.name(), FieldType.INT64, fields);
                 schema.getFields().add(new Field(0, FieldName.OFFSET.getKey(), SchemaBuilder.string().build()));
 
                 JSONObject jsonObject = new JSONObject();
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
@@ -153,7 +153,7 @@ public class RmqMetaReplicator extends SourceConnector {
             e.printStackTrace();
         }
 
-        return Utils.groupPartitions(new ArrayList<>(this.knownGroups), this.replicatorConfig.getTaskParallelism(), replicatorConfig);
+        return Utils.groupPartitions(new ArrayList<>(this.knownGroups), this.replicatorConfig.getTaskParallelism(), replicatorConfig, maxTasks);
     }
 
     private void prepare() {
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
@@ -153,7 +153,7 @@ public class RmqMetaReplicator extends SourceConnector {
             e.printStackTrace();
         }
 
-        return Utils.groupPartitions(new ArrayList<>(this.knownGroups), this.replicatorConfig.getTaskParallelism(), replicatorConfig, maxTasks);
+        return Utils.groupPartitions(new ArrayList<>(this.knownGroups), replicatorConfig, maxTasks);
     }
 
     private void prepare() {
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
@@ -45,6 +45,7 @@ import org.apache.rocketmq.remoting.exception.RemotingException;
 import org.apache.rocketmq.remoting.exception.RemotingSendRequestException;
 import org.apache.rocketmq.remoting.exception.RemotingTimeoutException;
 import org.apache.rocketmq.replicator.common.Utils;
+import org.apache.rocketmq.replicator.config.ConfigDefine;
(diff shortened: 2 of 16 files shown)


connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -117,38 +128,19 @@ public void start(KeyValue config) {
...
 
             for (Map.Entry<MessageQueue, OffsetWrapper> offsetTable : stats.getOffsetTable().entrySet()) {
-
                 MessageQueue mq = offsetTable.getKey();
                 long srcOffset = offsetTable.getValue().getConsumerOffset();
                 long targetOffset = this.store.convertTargetOffset(mq, group, srcOffset);
 
+                List<Field> fields = new ArrayList<Field>();
+                Schema schema = new Schema(SchemaEnum.OFFSET.name(), FieldType.STRING, fields);
+                schema.getFields().add(new Field(0, FieldName.OFFSET.getKey(), SchemaBuilder.string().build()));
```
odbozhou, line 137: offset use INT64  may be better

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
```diff
@@ -87,20 +88,25 @@ public RmqMetaReplicator() {
         executor = Executors.newSingleThreadScheduledExecutor(new BasicThreadFactory.Builder().namingPattern("RmqMetaReplicator-SourceWatcher-%d").daemon(true).build());
     }
 
-    @Override public String verifyAndSetConfig(KeyValue config) {
+    @Override public void validate(KeyValue config) {
         log.info("verifyAndSetConfig...");
         try {
             replicatorConfig.validate(config);
+            this.configValid = true;
         } catch (IllegalArgumentException e) {
```
odbozhou, line 96: I think the exception can be thrown upward, because there is no return, the upper layer cannot determine whether the verification is passed, otherwise the exception will be replenished, and there is no exception log, it is difficult to diagnose the problem

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
```diff
@@ -87,20 +88,25 @@ public RmqMetaReplicator() {
...
+            return;
         }
         this.prepare();
-        this.configValid = true;
-        return "";
+        return;
+    }
+
+    @Override public void init(KeyValue config) {
+
```
odbozhou, line 104: config should be set by init

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
```diff
@@ -89,26 +90,27 @@ private synchronized void startMQAdminTools() throws MQClientException {
...
 
     @Override
-    public String verifyAndSetConfig(KeyValue config) {
+    public void validate(KeyValue config) {
 
         // Check the need key.
         for (String requestKey : ConfigDefine.REQUEST_CONFIG) {
             if (!config.containsKey(requestKey)) {
-                return "Request config key: " + requestKey;
+                return;
```
odbozhou, line 98: Missing configuration should throw an exception instead of returning directly. Returning directly indicates that the verification has passed

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
```diff
@@ -89,26 +90,27 @@ private synchronized void startMQAdminTools() throws MQClientException {
...
             }
         }
 
         try {
             this.replicatorConfig.validate(config);
         } catch (IllegalArgumentException e) {
-            return e.getMessage();
+            return;
```
odbozhou, line 105: Exception should be throw

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
```diff
@@ -117,10 +119,10 @@ public void start() {
         }
 
         buildRoute();
-        startListner();
+        startListener();
```
odbozhou, line 122: startListener may not be needed, because the runtime has already done the load balancing logic, and put the load balancing queue in the context of each task

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
```diff
@@ -164,6 +166,11 @@ public boolean compare(Map<String, Set<TaskTopicInfo>> origin, Map<String, Set<T
         return true;
     }
 
+
+    @Override public void init(KeyValue config) {
+
```
odbozhou, line 171: config should be init by init

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
```diff
@@ -188,7 +195,7 @@ public Class<? extends Task> taskClass() {
     }
 
     @Override
-    public List<KeyValue> taskConfigs() {
+    public List<KeyValue> taskConfigs(int maxTasks) {
         if (!configValid) {
```
odbozhou, line 199: maxTasks can specify the number of tasks. Can you consider whether it needs to be divided into multiple tasks?

Slideee: For replicators, taskParallelism is used to control the number of tasks. The taskParallelism field and maxTask are semantically identical, so there is no definition for maxTask.

Slideee: > maxTasks can specify the number of tasks. Can you consider whether it needs to be divided into multiple tasks?

For replicators, taskParallelism is used to control the number of tasks. The taskParallelism field and maxTask are semantically identical, so there is no definition for maxTask.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
Slideee: > startListener may not be needed, because the runtime has already done the load balancing logic, and put the load balancing queue in the context of each task

I don't think it should be deleted, because the startListener method maintains changes in topic information, and the runtime maintains changes in task configuration and load balancing. Changes to a topic should not be maintained by the runtime :)

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
Slideee: > offset use INT64 may be better

done

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
Slideee: > I think the exception can be thrown upward, because there is no return, the upper layer cannot determine whether the verification is passed, otherwise the exception will be replenished, and there is no exception log, it is difficult to diagnose the problem

done

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
Slideee: > config should be set by init

For the Replicator, the validate method is already config initialized. So it's not initialized in the init method.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
Slideee: > Missing configuration should throw an exception instead of returning directly. Returning directly indicates that the verification has passed

done

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
Slideee: > Exception should be throw

done

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
Slideee: > config should be init by init

For the Replicator, the validate method is already config initialized. So it's not initialized in the init method.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
odbozhou: I made a mistake. The source connecotr should be needed to sense the change of the source mq topic. It is not necessary on the sink side.

Slideee: > I made a mistake. The source connecotr should be needed to sense the change of the source mq topic. It is not necessary on the sink side.

this method is in the source connector, what do you mean by sink side? in addition, can you review the latest commits :)

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceReplicator.java
odbozhou: The limitation of the previous version of the connect api, if the replicator wants to provide the ability of multiple tasks, only the replicator itself can provide the number of tasks, but we found that the ability of multiple tasks is general, and the replicator should not implement its own definition configuration, almost every specific connector implementation There is such a demand, so the specified configuration of the number of tasks should be unified. It is easier to use the maxTasks provided by the api to unify this capability. The connector shards the tasks according to the specific storage sharding. The actual number of task shards should be, For example, by topic, min (the number of tasks divided by topic, maxTasks) is configured.

Slideee: > The limitation of the previous version of the connect api, if the replicator wants to provide the ability of multiple tasks, only the replicator itself can provide the number of tasks, but we found that the ability of multiple tasks is general, and the replicator should not implement its own definition configuration, almost every specific connector implementation There is such a demand, so the specified configuration of the number of tasks should be unified. It is easier to use the maxTasks provided by the api to unify this capability. The connector shards the tasks according to the specific storage sharding. The actual number of task shards should be, For example, by topic, min (the number of tasks divided by topic, maxTasks) is configured.

Then I think the taskParallelism parameter should be removed and use maxTask instead of taskParallelism to control the number of tasks. Otherwise, writing like min (the number of tasks divided by topic, maxTasks) will only increase the cost of understanding, what do you think?

odbozhou: > 

1、Yes，maxTask should replace taskParallelism.
2、min (number of shards, maxTasks) should also be implemented. This is also easier to understand. If there are only 2 shards and maxTasks=5, the number of tasks can only be 2, which is why the number of tasks is maxTasks instead of taskNum. For example, the data source has only 2 topics, or 2 tables, maxTasks=5, the dimension of the shard is topic or table, and the final task data is 2, min(topicNum, maxTasks)

Slideee: > 

I understand what you mean, but the current replicator uses taskParallelism to do sharding. According to what you said, if 2 topics and taskParallelism is 1, then both topics are on 1 task (not 2 tasks). In this way, taskParallelism may be retained, then the logic is taskParallelism = min(taskParallelism, maxTasks)

odbozhou: There is no problem with multiple topics in one task. The main problem is that the configuration of the number of tasks can be unified. Basically, all connectors need this configuration.

Slideee: > There is no problem with multiple topics in one task. The main problem is that the configuration of the number of tasks can be unified. Basically, all connectors need this configuration.

I tried to submit a version, can you review it?

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/strategy/DivideTaskByTopic.java
```diff
@@ -28,10 +28,12 @@
...
 public class DivideTaskByTopic extends TaskDivideStrategy {
 
-    @Override public List<KeyValue> divide(Map<String, Set<TaskTopicInfo>> topicRouteMap, TaskDivideConfig tdc) {
+    @Override
+    public List<KeyValue> divide(Map<String, Set<TaskTopicInfo>> topicRouteMap, TaskDivideConfig tdc,
+        int maxTasks) {
 
         List<KeyValue> config = new ArrayList<KeyValue>();
-        int parallelism = tdc.getTaskParallelism();
+        int parallelism = Math.min(tdc.getTaskParallelism(), maxTasks);
```
odbozhou, line 36: 1、may be ConfigDefine.CONN_TASK_PARALLELISM  should  deprecate
   this.taskParallelism = config.getInt(ConfigDefine.CONN_TASK_PARALLELISM, 1);
2、if devide by topic
int parallelism = Math.min(topicNum, maxTasks);

Slideee: > ```
>  1、may be ConfigDefine.CONN_TASK_PARALLELISM  should  deprecate
> ```
> 
> this.taskParallelism = config.getInt(ConfigDefine.CONN_TASK_PARALLELISM, 1); 2、if devide by topic int parallelism = Math.min(topicNum, maxTasks);

as you mean 
if divide by queue, int parallelism = Math.min(queueNum, maxTasks) 
if divide by group, int parallelism = Math.min(groupNum, maxTasks) 
Is that right？
then the number of hash nodes is still compared with taskParallelism and maxTasks?

Slideee: > ```
>  1、may be ConfigDefine.CONN_TASK_PARALLELISM  should  deprecate
> ```
> 
> this.taskParallelism = config.getInt(ConfigDefine.CONN_TASK_PARALLELISM, 1); 2、if devide by topic int parallelism = Math.min(topicNum, maxTasks);

done

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/config/RmqConnectorConfig.java
```diff
@@ -25,7 +25,7 @@
 import org.apache.rocketmq.replicator.strategy.TaskDivideStrategy;
 
 public class RmqConnectorConfig {
-
+    @Deprecated
     private int taskParallelism;
```
odbozhou, line 29: Is it possible to delete this property directly?

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/config/RmqConnectorConfig.java
```diff
@@ -48,7 +48,7 @@ public class RmqConnectorConfig {
     public RmqConnectorConfig() {
     }
 
-    public void validate(KeyValue config) {
+    public void init(KeyValue config) {
         this.taskParallelism = config.getInt(ConfigDefine.CONN_TASK_PARALLELISM, 1);
```
odbozhou, line 52: Is it possible to delete this property directly?

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/config/TaskDivideConfig.java
```diff
@@ -28,6 +28,7 @@ public class TaskDivideConfig {
 
     private int dataType;
 
+    @Deprecated
     private int taskParallelism;
```
odbozhou, line 32: Is it possible to delete this property directly?

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/strategy/DivideTaskByQueue.java
```diff
@@ -31,10 +31,11 @@
 
 public class DivideTaskByQueue extends TaskDivideStrategy {
 
-    @Override public List<KeyValue> divide(Map<String, Set<TaskTopicInfo>> topicRouteMap, TaskDivideConfig tdc) {
+    @Override
+    public List<KeyValue> divide(Map<String, Set<TaskTopicInfo>> topicRouteMap, TaskDivideConfig tdc, int maxTasks) {
 
         List<KeyValue> config = new ArrayList<KeyValue>();
-        int parallelism = tdc.getTaskParallelism();
+        int parallelism = Math.min(tdc.getTaskParallelism(), maxTasks);
```
odbozhou, line 38: Is it possible to delete taskParallelism directly?

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqMetaReplicator.java
```diff
@@ -139,7 +153,7 @@ public List<KeyValue> taskConfigs() {
             e.printStackTrace();
         }
 
-        return Utils.groupPartitions(new ArrayList<>(this.knownGroups), this.replicatorConfig.getTaskParallelism(), replicatorConfig);
+        return Utils.groupPartitions(new ArrayList<>(this.knownGroups), this.replicatorConfig.getTaskParallelism(), replicatorConfig, maxTasks);
```
odbozhou, line 156: Is it possible to delete  taskParallelism directly?

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/strategy/DivideTaskByConsistentHash.java
```diff
@@ -32,10 +32,11 @@
 import org.apache.rocketmq.replicator.config.TaskTopicInfo;
 
 public class DivideTaskByConsistentHash extends TaskDivideStrategy {
-    @Override public List<KeyValue> divide(Map<String, Set<TaskTopicInfo>> topicMap, TaskDivideConfig tdc) {
+    @Override
+    public List<KeyValue> divide(Map<String, Set<TaskTopicInfo>> topicMap, TaskDivideConfig tdc, int maxTasks) {
 
         List<KeyValue> config = new ArrayList<>();
-        int parallelism = tdc.getTaskParallelism();
+        int parallelism = Math.min(tdc.getTaskParallelism(), maxTasks);
```
odbozhou, line 39: Is it possible to delete taskParallelism directly?

Slideee: > Is it possible to delete taskParallelism directly?

if divide by hash or divide by queue or divide by group，directly use the maxTasks parameter to control the number of tasks?

odbozhou (approved):

Issue #132, opened 2022-05-11
rocketmq-replicator optimization


zhangjidi2016:
After rocketmq-replicator adapt to the new conenct api merge, the pull code is run and the following problems are found:

Fixed:
1. The message content is inconsistent. The message content after synchronization is {"MessageExt": "original message content "} 
2. Message attributes are not synchronized

Function:
1. When synchronizing messages from the source cluster to the target cluster, we can define filtering rules to synchronize only the filtered messages

Optimization:
1. Keep replicator and Runtime versions of some of the same dependencies the same to prevent conflicts

replicator适配了新的connect api合并后，拉取代码运行了下，主要发现以下几点问题：

修复：
1.消息同步内容不一致，同步后的消息内容为{"MessageExt": "原消息内容"}
2.消息属性没有同步

功能：
1.源和目标集群消息同步时，可以定义过滤规则，只同步过滤后的消息

优化：
1.保持replicator和runtime中的一些相同依赖包版本相同，防止冲突

PR #136, opened 2022-05-12
[ISSUE #132]Define filtering rules to synchronize only filtered messages

zhangjidi2016:

#132

Utilize RocketMQ's Tag filtering mechanism，before message synchronization, define filtering rules and synchronize only the messages that meet the filtering rules.

When sending an http request to start the rocketmq-replicator, add the filter-rule parameter to the parameter list
image


odbozhou:
If there are multiple topics, does the tag of each topic have to be the same?

zhangjidi2016:
> If there are multiple topics, does the tag of each topic have to be the same?

Yes, all topics in this request will use this filtering rule (tag)

odbozhou:
> 

Is it possible to implement different tags for different topics?

RockteMQ-AI (approved):
LGTM. Changes look good.

---

PR #145, opened 2022-05-24
rocketmq-replicator 同步消息异常

LittleBoy18:
1、readme 指导操作有误
2、同步消息到白名单设置的topic

### DIFF
--- a/connectors/rocketmq-replicator/README.md
+++ b/connectors/rocketmq-replicator/README.md
@@ -28,7 +28,7 @@ mvn clean install -Prelease-all -DskipTest -U
 同步topic和消息
 ````
 http://${runtime-ip}:${runtime-port}/connectors/${rocketmq-replicator-name}
-?config={"connector-class":"org.apache.rocketmq.replicator.RmqSourceReplicator","source-rocketmq":"xxxx:9876","target-rocketmq":"xxxxxxx:9876","replicator-store-topic":"replicatorTopic","taskDivideStrategy":"0","white-list":"TopicTest,TopicTest2","task-parallelism":"2","source-record-converter":"org.apache.rocketmq.connect.runtime.converter.JsonConverter"}
+?config={"connector-class":"org.apache.rocketmq.replicator.RmqSourceReplicator","source-rocketmq":"xxxx:9876","target-rocketmq":"xxxxxxx:9876","replicator-store-topic":"replicatorTopic","taskDivideStrategy":"0","white-list":"TopicTest,TopicTest2","task-parallelism":"2","source-record-converter":"org.apache.rocketmq.connect.runtime.converter.RocketMQConverter"}
 ````
 
 
--- a/connectors/rocketmq-replicator/README.md
+++ b/connectors/rocketmq-replicator/README.md
@@ -44,7 +44,7 @@ http://${runtime-ip}:${runtime-port}/connectors/${rocketmq-replicator-name}/stop
 注：此功能尚不成熟还需要后续版本优化
 ````
 http://${runtime-ip}:${runtime-port}/connectors/${rocketmq-replicator-name}
-?config={"connector-class":"org.apache.rocketmq.replicator.RmqMetaReplicator","source-rocketmq":"xxxx:9876","target-rocketmq":"xxxxxxx:9876","replicator-store-topic":"replicatorTopic","offset.sync.topic":"syncTopic","taskDivideStrategy":"0","white-list":"TopicTest,TopicTest2","task-parallelism":"2","source-record-converter":"org.apache.rocketmq.connect.runtime.converter.JsonConverter"}
+?config={"connector-class":"org.apache.rocketmq.replicator.RmqMetaReplicator","source-rocketmq":"xxxx:9876","target-rocketmq":"xxxxxxx:9876","target-cluster":"test1-rocketmq","source-cluster":"test1-rocketmq","replicator-store-topic":"replicatorTopic","offset.sync.topic":"syncTopic","taskDivideStrategy":"0","white-list":"TestGroup","task-parallelism":"2","source-record-converter":"org.apache.rocketmq.connect.runtime.converter.RocketMQMetaConverter"}
 ````
 
 
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
@@ -30,11 +30,16 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.HashMap;
+import java.util.Set;
+import java.util.Iterator;
 import java.util.concurrent.TimeUnit;
 import org.apache.rocketmq.client.exception.MQClientException;
 import org.apache.rocketmq.common.admin.ConsumeStats;
 import org.apache.rocketmq.common.admin.OffsetWrapper;
 import org.apache.rocketmq.common.message.MessageQueue;
+import org.apache.rocketmq.common.protocol.body.ClusterInfo;
+import org.apache.rocketmq.common.protocol.route.BrokerData;
 import org.apache.rocketmq.replicator.common.Utils;
 import org.apache.rocketmq.replicator.config.ConfigUtil;
 import org.apache.rocketmq.replicator.config.TaskConfig;
@@ -52,6 +57,7 @@ public class MetaSourceTask extends SourceTask {
     private final String taskId;
     private final TaskConfig config;
     private DefaultMQAdminExt srcMQAdminExt;
+    private DefaultMQAdminExt tarMQAdminExt;
     private volatile boolean started = false;
 
     private OffsetSyncStore store;
@@ -77,6 +83,7 @@ public class MetaSourceTask extends SourceTask {
 
         try {
             this.srcMQAdminExt = Utils.startMQAdminTool(this.config);
+            this.tarMQAdminExt = Utils.startTarMQAdminTool(this.config);
         } catch (MQClientException e) {
             log.error("Replicator task start failed for `startMQAdminTool` exception.", e);
             throw new IllegalStateException("Replicator task start failed for `startMQAdminTool` exception.");
@@ -92,6 +99,7 @@ public class MetaSourceTask extends SourceTask {
             started = false;
         }
         srcMQAdminExt.shutdown();
+        tarMQAdminExt.shutdown();
     }
 
     @Override
@@ -120,27 +128,37 @@ public class MetaSourceTask extends SourceTask {
         List<ConnectRecord> res = new ArrayList<>();
         for (String group : groups) {
             ConsumeStats stats;
+            String brokerAddresMaster="";
+            String brokerName="";
             try {
                 stats = this.srcMQAdminExt.examineConsumeStats(group);
+                ClusterInfo clusterInfo = this.tarMQAdminExt.examineBrokerClusterInfo();
+                HashMap<String, Set<String>> clusterAddrTable = clusterInfo.getClusterAddrTable();
+                HashMap<String, BrokerData> brokerAddrTable = clusterInfo.getBrokerAddrTable();
+                Set<String> clusterNameSet = clusterAddrTable.get(this.config.getTargetCluster());
+                Iterator<String> it = clusterNameSet.iterator();
+                while (it.hasNext()){
+                   String clusterName = it.next();
+                   BrokerData brokerData = brokerAddrTable.get(clusterName);
+                   HashMap<Long, String> brokerAddrs = brokerData.getBrokerAddrs();
+                   brokerAddresMaster = brokerAddrs.get(new Long(0));
+                   brokerName = brokerData.getBrokerName();
+                    for (Map.Entry<MessageQueue, OffsetWrapper> offsetTable : stats.getOffsetTable().entrySet()) {
+                        MessageQueue mq = offsetTable.getKey();
+                        long srcOffset = offsetTable.getValue().getConsumerOffset();
+                        long targetOffset = this.store.convertTargetOffset(mq, group, srcOffset);
+                        try{
+                           if (brokerName.equals(mq.getBrokerName())){
+                               this.tarMQAdminExt.updateConsumeOffset(brokerAddresMaster,group,mq,targetOffset);
+                           }
+                        }catch (Exception e){
+                            log.error("admin update consumer offset err", e);
+                        }
+                    }
+                }
             } catch (Exception e) {
                 log.error("admin get consumer info failed for consumer groups: " + group, e);
                 continue;
-            }
-
-            for (Map.Entry<MessageQueue, OffsetWrapper> offsetTable : stats.getOffsetTable().entrySet()) {
-                MessageQueue mq = offsetTable.getKey();
-                long srcOffset = offsetTable.getValue().getConsumerOffset();
-                long targetOffset = this.store.convertTargetOffset(mq, group, srcOffset);
-
-                List<Field> fields = new ArrayList<Field>();
-                Schema schema = new Schema(SchemaEnum.OFFSET.name(), FieldType.INT64, fields);
-                schema.getFields().add(new Field(0, FieldName.OFFSET.getKey(), SchemaBuilder.string().build()));
-
-                JSONObject jsonObject = new JSONObject();
-                jsonObject.put(FieldName.OFFSET.getKey(), targetOffset);
-                ConnectRecord connectRecord = new ConnectRecord(Utils.offsetKey(mq),
-                    Utils.offsetValue(srcOffset), System.currentTimeMillis(), schema, jsonObject.toJSONString());
-                res.add(connectRecord);
             }
         }
         return res;
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceTask.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceTask.java
@@ -19,6 +19,7 @@ package org.apache.rocketmq.replicator;
 import com.alibaba.fastjson.JSON;
 import com.alibaba.fastjson.JSONObject;
 import io.openmessaging.KeyValue;
+import io.openmessaging.internal.DefaultKeyValue;
 import io.openmessaging.connector.api.component.task.source.SourceTask;
 import io.openmessaging.connector.api.component.task.source.SourceTaskContext;
 import io.openmessaging.connector.api.data.ConnectRecord;
@@ -179,6 +180,10 @@ public class RmqSourceTask extends SourceTask {
                                 final Set<String> keys = properties.keySet();
                                 keys.forEach(key -> connectRecord.addExtension(key, properties.get(key)));
                                 connectRecord.addExtension("topic",taskTopicConfig.getTargetTopic());
+                                connectRecord.addExtension("brokerName",msg.getBrokerName());
+                                KeyValue kv = new DefaultKeyValue();
+                                kv.put("queueId",msg.getQueueId());
+                                connectRecord.addExtension(kv);
                                 res.add(connectRecord);
                             }
                             break;
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceTask.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceTask.java
@@ -178,6 +178,7 @@ public class RmqSourceTask extends SourceTask {
                                 final Map<String, String> properties = msg.getProperties();
                                 final Set<String> keys = properties.keySet();
                                 keys.forEach(key -> connectRecord.addExtension(key, properties.get(key)));
+                                connectRecord.addExtension("topic",taskTopicConfig.getTargetTopic());
                                 res.add(connectRecord);
                             }
                             break;
--- a/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/common/Utils.java
+++ b/connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/common/Utils.java
@@ -145,7 +145,9 @@ public class Utils {
                 assigned++;
             }
             keyValue.put(TaskConfigEnum.TASK_STORE_ROCKETMQ.getKey(), tdc.getStoreTopic());
+            keyValue.put(TaskConfigEnum.TASK_TARGET_ROCKETMQ.getKey(), tdc.getTargetNamesrvs());
             keyValue.put(TaskConfigEnum.TASK_SOURCE_ROCKETMQ.getKey(), tdc.getSrcNamesrvs());
+            keyValue.put(TaskConfigEnum.TASK_TARGET_CLUSTER.getKey(), tdc.getTargetCluster());
             keyValue.put(TaskConfigEnum.TASK_SOURCE_CLUSTER.getKey(), tdc.getSrcCluster());
             keyValue.put(TaskConfigEnum.TASK_OFFSET_SYNC_TOPIC.getKey(), tdc.getOffsetSyncTopic());
             keyValue.put(TaskConfigEnum.TASK_DATA_TYPE.getKey(), DataType.OFFSET.ordinal());
@@ -198,10 +200,26 @@ public class Utils {
         if (taskConfig.isSrcAclEnable()) {
             rpcHook = new AclClientRPCHook(new SessionCredentials(taskConfig.getSrcAccessKey(), taskConfig.getSrcSecretKey()));
         }
+        DefaultMQAdminExt sourceMQAdminExt = new DefaultMQAdminExt(rpcHook);
+        sourceMQAdminExt.setNamesrvAddr(taskConfig.getSourceRocketmq());
+        sourceMQAdminExt.setAdminExtGroup(ConstDefine.REPLICATOR_TASK_ADMIN_GROUP);
+        sourceMQAdminExt.setInstanceName(Utils.createUniqInstanceName(taskConfig.getSourceRocketmq()));
+
+        sourceMQAdminExt.start();
+        log.info("Source: RocketMQ sourceMQAdminExt started.");
+
+        return sourceMQAdminExt;
+    }
+
+    public static DefaultMQAdminExt startTarMQAdminTool(TaskConfig taskConfig) throws MQClientException {
+        RPCHook rpcHook = null;
+        if (taskConfig.isSrcAclEnable()) {
+            rpcHook = new AclClientRPCHook(new SessionCredentials(taskConfig.getSrcAccessKey(), taskConfig.getSrcSecretKey()));
+        }
         DefaultMQAdminExt targetMQAdminExt = new DefaultMQAdminExt(rpcHook);
-        targetMQAdminExt.setNamesrvAddr(taskConfig.getSourceRocketmq());
+        targetMQAdminExt.setNamesrvAddr(taskConfig.getTargetRocketmq());
         targetMQAdminExt.setAdminExtGroup(ConstDefine.REPLICATOR_TASK_ADMIN_GROUP);
-        targetMQAdminExt.setInstanceName(Utils.createUniqInstanceName(taskConfig.getSourceRocketmq()));
+        targetMQAdminExt.setInstanceName(Utils.createUniqInstanceName(taskConfig.getTargetRocketmq()));
 
         targetMQAdminExt.start();
         log.info("TARGET: RocketMQ targetMQAdminExt started.");
--- a/rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
+++ b/rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
@@ -307,6 +307,10 @@ public class WorkerSourceTask implements WorkerTask {
                         log.error("Send record, message size is greater than {} bytes, sourceDataEntry: {}", RuntimeConfigDefine.MAX_MESSAGE_SIZE, JSON.toJSONString(sourceDataEntry));
                         continue;
                     }
+                    String targetTopic = sourceDataEntry.getExtension("topic");
+                    if (targetTopic != null){
+                        sourceMessage.setTopic(targetTopic);
+                    }
                     sourceMessage.setBody(messageBody);
                 }
             } else {
--- a/rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
+++ b/rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
@@ -46,10 +46,12 @@ import org.apache.rocketmq.client.producer.SendCallback;
 import org.apache.rocketmq.common.message.Message;
 import org.apache.rocketmq.common.message.MessageAccessor;
 import org.apache.rocketmq.common.message.MessageConst;
+import org.apache.rocketmq.common.message.MessageQueue;
 import org.apache.rocketmq.connect.runtime.common.ConnectKeyValue;
 import org.apache.rocketmq.connect.runtime.common.LoggerName;
 import org.apache.rocketmq.connect.runtime.config.RuntimeConfigDefine;
 import org.apache.rocketmq.connect.runtime.converter.RocketMQConverter;
+import org.apache.rocketmq.connect.runtime.converter.RocketMQMetaConverter;
 import org.apache.rocketmq.connect.runtime.service.PositionManagementService;
 import org.apache.rocketmq.connect.runtime.stats.ConnectStatsManager;
 import org.apache.rocketmq.connect.runtime.stats.ConnectStatsService;
@@ -265,6 +267,9 @@ public class WorkerSourceTask implements WorkerTask {
      */
     private void sendRecord() throws InterruptedException, RemotingException, MQClientException {
         for (ConnectRecord sourceDataEntry : toSendRecord) {
+            if (recordConverter instanceof RocketMQMetaConverter){
+                return;
+            }
             RecordPosition position = sourceDataEntry.getPosition();
             RecordOffset offset = position.getOffset();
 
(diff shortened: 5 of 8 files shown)


LittleBoy18:
1、优化了 消息同步
2、优化了 消费进度同步

tsunghanjacktsai:
Instead of RmqSourceTask and MetaSourceTask, how about trying the newer version of rocketmq-replicator, which applies the ReplicatorSourceConnector and ReplicatorSourceTask and provides more complete features?

RockteMQ-AI:
⚠️ **Merge conflict detected**

This PR has conflicts with the base branch and cannot be merged. Please rebase or merge the base branch into your branch and resolve the conflicts:

```bash
git fetch origin
git checkout fix-syncmessage
git rebase origin/main
# resolve conflicts, then:
git push --force-with-lease
```

This is a one-time reminder. Feel free to @mention me for a re-review after conflicts are resolved.

---
*Automated notification by github-manager-bot*

RockteMQ-AI (commented):
### Summary
This PR modifies 8 file(s) with 455 lines of diff. No test changes detected — consider adding test coverage.

---

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -30,11 +30,16 @@
 import java.util.Collections;
```
RockteMQ-AI, line 30: No test changes detected alongside source modifications. Consider adding tests to cover the changes.

RockteMQ-AI:
This PR has been open since May 2022 (over 4 years) and currently has merge conflicts with the master branch.

**Status check:** Is this replicator sync fix still relevant? The changes to MetaSourceTask, RmqSourceTask, and WorkerSourceTask appear to address real issues, but the code needs to be rebased.

If this is still being worked on, please rebase and update. If abandoned, consider closing.

---

RockteMQ-AI (commented):
### Summary
PR received and logged for review. This PR requires detailed code review by a maintainer.

**Diff size:** 455 lines
**Author:** LittleBoy18 (CONTRIBUTOR)

---
*Automated review by RockteMQ-AI*

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
...
         for (String group : groups) {
             ConsumeStats stats;
+            String brokerAddresMaster="";
+            String brokerName="";
             try {
                 stats = this.srcMQAdminExt.examineConsumeStats(group);
+                ClusterInfo clusterInfo = this.tarMQAdminExt.examineBrokerClusterInfo();
+                HashMap<String, Set<String>> clusterAddrTable = clusterInfo.getClusterAddrTable();
+                HashMap<String, BrokerData> brokerAddrTable = clusterInfo.getBrokerAddrTable();
+                Set<String> clusterNameSet = clusterAddrTable.get(this.config.getTargetCluster());
```
RockteMQ-AI, line 138: NPE risk: `clusterAddrTable.get(this.config.getTargetCluster())` can return null if the target cluster name is misconfigured or not yet registered in the target cluster topology. The subsequent `clusterNameSet.iterator()` would throw NullPointerException with no useful error message. Add a null check and log a meaningful error identifying the missing cluster.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
...
+                Iterator<String> it = clusterNameSet.iterator();
+                while (it.hasNext()){
+                   String clusterName = it.next();
+                   BrokerData brokerData = brokerAddrTable.get(clusterName);
```
RockteMQ-AI, line 142: NPE risk: `brokerAddrTable.get(clusterName)` may return null if a broker listed in the cluster table is not yet in the broker address table (race during broker add/remove). The subsequent `brokerData.getBrokerAddrs()` would NPE. Additionally, `brokerAddrs.get(new Long(0))` may return null if no master (brokerId=0) is available, causing `updateConsumeOffset(null, ...)` to silently fail inside the catch block. Add null checks for brokerData, brokerAddrs, and brokerAddresMaster.

rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
```diff
@@ -307,7 +312,53 @@ private void sendRecord() throws InterruptedException, RemotingException, MQClie
                         log.error("Send record, message size is greater than {} bytes, sourceDataEntry: {}", RuntimeConfigDefine.MAX_MESSAGE_SIZE, JSON.toJSONString(sourceDataEntry));
                         continue;
                     }
+                    String targetTopic = sourceDataEntry.getExtension("topic");
+                    if (targetTopic != null){
+                        sourceMessage.setTopic(targetTopic);
+                    }
                     sourceMessage.setBody(messageBody);
+                    int queueId = sourceDataEntry.getExtensions().getInt("queueId");
```
RockteMQ-AI, line 320: NPE risk: `sourceDataEntry.getExtensions().getInt("queueId")` assumes getExtensions() is non-null and contains the "queueId" key. For any record not produced by RmqSourceTask (which is the only place queueId is set), this throws NPE or NoSuchElementException. Should null-check getExtensions() and provide a safe default or skip the MessageQueue-specific send path.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
         List<ConnectRecord> res = new ArrayList<>();
         for (String group : groups) {
             ConsumeStats stats;
+            String brokerAddresMaster="";
+            String brokerName="";
             try {
                 stats = this.srcMQAdminExt.examineConsumeStats(group);
+                ClusterInfo clusterInfo = this.tarMQAdminExt.examineBrokerClusterInfo();
```
RockteMQ-AI, line 135: Performance: `this.tarMQAdminExt.examineBrokerClusterInfo()` is called inside the per-group loop. This is an expensive RPC that returns the same target cluster topology for every consumer group. Hoist this call outside the for-loop and reuse the result across all groups.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -92,6 +99,7 @@ public void stop() {
             started = false;
         }
         srcMQAdminExt.shutdown();
+        tarMQAdminExt.shutdown();
     }
```
RockteMQ-AI, line 104: NPE risk in stop(): `tarMQAdminExt.shutdown()` is called unconditionally. If start() fails after creating srcMQAdminExt but before assigning tarMQAdminExt (e.g., startTarMQAdminTool throws MQClientException), stop() will throw NPE on a null tarMQAdminExt. Should null-check before calling shutdown.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
...
+            String brokerAddresMaster="";
+            String brokerName="";
             try {
                 stats = this.srcMQAdminExt.examineConsumeStats(group);
+                ClusterInfo clusterInfo = this.tarMQAdminExt.examineBrokerClusterInfo();
+                HashMap<String, Set<String>> clusterAddrTable = clusterInfo.getClusterAddrTable();
+                HashMap<String, BrokerData> brokerAddrTable = clusterInfo.getBrokerAddrTable();
+                Set<String> clusterNameSet = clusterAddrTable.get(this.config.getTargetCluster());
+                Iterator<String> it = clusterNameSet.iterator();
+                while (it.hasNext()){
```
RockteMQ-AI, line 140: Performance: The while-loop iterates all brokers in the target cluster, and for each broker iterates ALL message queues from the consume stats, only updating the offset when `brokerName.equals(mq.getBrokerName())`. This is O(brokers x queues). Build a Map<brokerName, brokerAddress> first, then iterate the message queues once and look up the matching broker address.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/common/Utils.java
```diff
@@ -194,14 +196,30 @@ public static DefaultMQAdminExt startTargetMQAdminTool(
...
+        sourceMQAdminExt.setAdminExtGroup(ConstDefine.REPLICATOR_TASK_ADMIN_GROUP);
+        sourceMQAdminExt.setInstanceName(Utils.createUniqInstanceName(taskConfig.getSourceRocketmq()));
+
+        sourceMQAdminExt.start();
+        log.info("Source: RocketMQ sourceMQAdminExt started.");
+
+        return sourceMQAdminExt;
+    }
+
+    public static DefaultMQAdminExt startTarMQAdminTool(TaskConfig taskConfig) throws MQClientException {
```
RockteMQ-AI, line 214: startTarMQAdminTool uses source ACL credentials (`isSrcAclEnable()`, `getSrcAccessKey()`, `getSrcSecretKey()`) for the target cluster admin tool. If the target cluster has ACL enabled with different credentials than the source, authentication will fail. TaskConfig should have separate target ACL config fields, or at minimum this limitation should be documented.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceTask.java
```diff
@@ -178,6 +179,11 @@ private List<ConnectRecord> pollCommonMessage() {
                                 final Map<String, String> properties = msg.getProperties();
                                 final Set<String> keys = properties.keySet();
                                 keys.forEach(key -> connectRecord.addExtension(key, properties.get(key)));
+                                connectRecord.addExtension("topic",taskTopicConfig.getTargetTopic());
+                                connectRecord.addExtension("brokerName",msg.getBrokerName());
+                                KeyValue kv = new DefaultKeyValue();
+                                kv.put("queueId",msg.getQueueId());
+                                connectRecord.addExtension(kv);
```
RockteMQ-AI, line 186: Compatibility: Uses `io.openmessaging.internal.DefaultKeyValue`, an internal package not part of the public OMS API. This may break in future dependency versions. Additionally, verify that `connectRecord.addExtension(KeyValue)` properly merges the int-typed queueId value and that `getExtensions().getInt("queueId")` in WorkerSourceTask can round-trip it, since topic and brokerName are added as String extensions but queueId is added via a KeyValue with an int.

rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
```diff
@@ -307,7 +312,53 @@ private void sendRecord() throws InterruptedException, RemotingException, MQClie
...
+                    String brokerName = sourceDataEntry.getExtension("brokerName");
+                    MessageQueue mq = new MessageQueue(targetTopic,brokerName,queueId);
```
RockteMQ-AI, line 322: Correctness: `new MessageQueue(targetTopic, brokerName, queueId)` is constructed unconditionally, but targetTopic and brokerName may be null — the null check at line 316 only guards `sourceMessage.setTopic()`, not the MessageQueue construction. Sending to a MessageQueue with null topic or broker name will fail at the broker. Validate these fields before entering the RocketMQConverter branch or ensure they are always set by the source task.

rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
```diff
@@ -265,6 +267,9 @@ public void cleanup() {
      */
     private void sendRecord() throws InterruptedException, RemotingException, MQClientException {
         for (ConnectRecord sourceDataEntry : toSendRecord) {
+            if (recordConverter instanceof RocketMQMetaConverter){
```
RockteMQ-AI, line 270: The early `return` (not `continue`) for RocketMQMetaConverter skips the `toSendRecord = null` cleanup at the end of sendRecord(). If the framework does not reset toSendRecord each cycle, records could accumulate across poll cycles for meta tasks. Use `continue` or ensure toSendRecord is cleared outside this method for meta tasks.

rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
```diff
@@ -307,7 +312,53 @@ private void sendRecord() throws InterruptedException, RemotingException, MQClie
                         log.error("Send record, message size is greater than {} bytes, sourceDataEntry: {}", RuntimeConfigDefine.MAX_MESSAGE_SIZE, JSON.toJSONString(sourceDataEntry));
                         continue;
                     }
+                    String targetTopic = sourceDataEntry.getExtension("topic");
```
RockteMQ-AI, line 315: Maintainability: The SendCallback implementation (onSuccess/onException with stats tracking and position storage) is duplicated verbatim between the RocketMQConverter branch and the else branch (~40 lines each). Extract this into a helper method to avoid divergence bugs when one copy is updated but not the other.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
...
-
-                JSONObject jsonObject = new JSONObject();
-                jsonObject.put(FieldName.OFFSET.getKey(), targetOffset);
-                ConnectRecord connectRecord = new ConnectRecord(Utils.offsetKey(mq),
-                    Utils.offsetValue(srcOffset), System.currentTimeMillis(), schema, jsonObject.toJSONString());
-                res.add(connectRecord);
-            }
         }
         return res;
     }
```
RockteMQ-AI, line 165: Offset tracking concern: poll() now always returns an empty ConnectRecord list, and sendRecord() returns early for RocketMQMetaConverter, bypassing positionStorageWriter entirely. The framework's position/offset tracking is completely skipped for meta tasks. On task restart or reassignment, no progress is recorded. Verify this is intentional — the direct updateConsumeOffset approach may need at-least-once delivery guarantees that the framework no longer provides for this path.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
         List<ConnectRecord> res = new ArrayList<>();
```
RockteMQ-AI, line 128: No test coverage for the substantially changed meta offset sync logic (direct updateConsumeOffset, new startTarMQAdminTool, ClusterInfo traversal) or the RocketMQConverter send-to-specific-queue path in WorkerSourceTask. These are significant behavioral changes that warrant integration tests, especially given the NPE risks identified above.

RockteMQ-AI (changes requested):
### Summary
Review of PR #145: rocketmq-replicator 同步消息异常

**Findings:** 13 issue(s) identified (3 critical).
**CLA:** unknown

---

rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
RockteMQ-AI, line 320: Queue routing is applied unconditionally to every record in the `null == recordConverter || recordConverter instanceof RocketMQConverter` branch. ConnectRecord.getExtensions() returns null unless addExtension() was called (putExtendMsgProperty in this same method null-checks it), so `sourceDataEntry.getExtensions().getInt("queueId")` throws NPE for any record without extensions. Even when extensions exist but lack queueId/brokerName/topic, DefaultKeyValue.getInt() returns 0 and brokerName/targetTopic are null, so `new MessageQueue(null, null, 0)` is passed to producer.send() and the send fails. This is shared runtime code: every existing source connector that sets no converter or uses RocketMQConverter (not just the replicator) breaks. Guard this block (only route to a MessageQueue when topic+brokerName+queueId are all present) and fall back to producer.send(sourceMessage) otherwise.

rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
RockteMQ-AI, line 270: The RocketMQMetaConverter early-return sits inside the per-record loop and skips the `toSendRecord = null` cleanup at the end of sendRecord(). In run(), poll() is only invoked when toSendRecord is empty, so any meta-converter task that returns at least one record would busy-spin forever (100% CPU, no further polls, no position commits). Today MetaSourceTask.poll() always returns an empty list so this is dead code, but the check is converter-scoped, not record-scoped — move it before the loop (or clear the list) so the invariant can't be broken by a future task that emits records with this converter.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceTask.java
```diff
@@ -178,6 +179,11 @@ private List<ConnectRecord> pollCommonMessage() {
                                 final Map<String, String> properties = msg.getProperties();
                                 final Set<String> keys = properties.keySet();
                                 keys.forEach(key -> connectRecord.addExtension(key, properties.get(key)));
+                                connectRecord.addExtension("topic",taskTopicConfig.getTargetTopic());
```
RockteMQ-AI, line 182: Adding routing metadata as plain extensions corrupts replicated messages: original message properties are copied to extensions first (line 181), then addExtension("topic"/"brokerName"/"queueId") overwrites any same-named user property, silently losing source data. Additionally, putExtendMsgProperty writes all extensions back as message properties (connect-ext-topic, connect-ext-brokerName, connect-ext-queueId), so every replicated message gains three properties the original never had. Use a dedicated prefix (e.g. connect-internal-*) or a separate mechanism for routing metadata instead of the generic "topic" key, which also risks hijacking the routing of any other connector whose records happen to carry a "topic" extension.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
...
             ConsumeStats stats;
+            String brokerAddresMaster="";
+            String brokerName="";
             try {
                 stats = this.srcMQAdminExt.examineConsumeStats(group);
+                ClusterInfo clusterInfo = this.tarMQAdminExt.examineBrokerClusterInfo();
+                HashMap<String, Set<String>> clusterAddrTable = clusterInfo.getClusterAddrTable();
+                HashMap<String, BrokerData> brokerAddrTable = clusterInfo.getBrokerAddrTable();
+                Set<String> clusterNameSet = clusterAddrTable.get(this.config.getTargetCluster());
+                Iterator<String> it = clusterNameSet.iterator();
```
RockteMQ-AI, line 139: `clusterAddrTable.get(this.config.getTargetCluster())` returns null when target-cluster is not configured (older connector configs; DefaultKeyValue.put even stores null as the string "null") and `clusterNameSet.iterator()` NPEs. The broad catch swallows it with the misleading message "admin get consumer info failed" and offset sync silently does nothing — an operational trap. Validate targetCluster/targetRocketmq in validate()/start() and fail fast, and null-check the lookup with an explicit error.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/common/Utils.java
```diff
@@ -194,14 +196,30 @@ public static DefaultMQAdminExt startTargetMQAdminTool(
...
         RPCHook rpcHook = null;
         if (taskConfig.isSrcAclEnable()) {
```
RockteMQ-AI, line 216: startTarMQAdminTool connects to taskConfig.getTargetRocketmq() but authenticates with SOURCE ACL credentials (isSrcAclEnable/getSrcAccessKey/getSrcSecretKey). TaskConfig has no target ACL fields even though RmqConnectorConfig does (see startTargetMQAdminTool using isTargetAclEnable), so with ACL enabled on the target cluster under different credentials every admin call fails. Also this method is a near-duplicate of startMQAdminTool/startTargetMQAdminTool — add target ACL fields to TaskConfig and reuse one implementation.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
...
+                   BrokerData brokerData = brokerAddrTable.get(clusterName);
+                   HashMap<Long, String> brokerAddrs = brokerData.getBrokerAddrs();
+                   brokerAddresMaster = brokerAddrs.get(new Long(0));
+                   brokerName = brokerData.getBrokerName();
+                    for (Map.Entry<MessageQueue, OffsetWrapper> offsetTable : stats.getOffsetTable().entrySet()) {
+                        MessageQueue mq = offsetTable.getKey();
+                        long srcOffset = offsetTable.getValue().getConsumerOffset();
+                        long targetOffset = this.store.convertTargetOffset(mq, group, srcOffset);
+                        try{
+                           if (brokerName.equals(mq.getBrokerName())){
```
RockteMQ-AI, line 151: Offset sync matches the TARGET broker name against the SOURCE MessageQueue broker name, and WorkerSourceTask similarly sends to the same queueId/brokerName — the whole change assumes the target cluster has identical broker names and at least as many queues per topic. When names or queue counts differ, offsets are silently skipped here and data sends throw MQClientException in the runtime (records dropped). Also brokerAddrs.get(new Long(0)) returns null when no master is registered, making updateConsumeOffset fail. Validate the target route exists and log a clear warning otherwise; use Long.valueOf(0)/MixAll.MASTER_ID instead of new Long(0).

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -92,6 +99,7 @@ public void stop() {
             started = false;
         }
         srcMQAdminExt.shutdown();
+        tarMQAdminExt.shutdown();
```
RockteMQ-AI, line 102: stop() calls tarMQAdminExt.shutdown() without a null guard. If start() throws after srcMQAdminExt started (e.g. startTarMQAdminTool failure), tarMQAdminExt is null and stop() NPEs, masking the real error, while the already-started srcMQAdminExt is never shut down (resource leak). Guard both shutdowns and release srcMQAdminExt when the target admin fails to start.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
RockteMQ-AI, line 135: examineBrokerClusterInfo() is an expensive admin RPC invoked inside the per-group loop on every poll, and poll() is called in a tight runtime loop with no sleep, so cluster info is re-fetched G times per iteration. Hoist the call outside the groups loop and cache it (or refresh periodically); the nested group × broker × queue iteration can also be inverted so each queue is matched against a broker-name map.

connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java
```diff
@@ -120,28 +128,38 @@ public void resume() {
...
-                schema.getFields().add(new Field(0, FieldName.OFFSET.getKey(), SchemaBuilder.string().build()));
-
-                JSONObject jsonObject = new JSONObject();
-                jsonObject.put(FieldName.OFFSET.getKey(), targetOffset);
-                ConnectRecord connectRecord = new ConnectRecord(Utils.offsetKey(mq),
-                    Utils.offsetValue(srcOffset), System.currentTimeMillis(), schema, jsonObject.toJSONString());
-                res.add(connectRecord);
-            }
         }
         return res;
```
RockteMQ-AI, line 164: poll() now always returns an empty list — the record/schema machinery (ConnectRecord, Schema, Field, SchemaBuilder, JSONObject, FieldName, SchemaEnum imports, and the `res` variable) is dead code and poll() is purely side-effecting. Note also that OffsetSyncStore.sync() is never invoked (its consumer is never started), so convertTargetOffset degenerates to identity and updateConsumeOffset copies source offsets verbatim — wrong if target offsets diverge (e.g. replication didn't start at offset 0). Clean up the dead code and document that offset.sync.topic is no longer used by this path.

rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java
```diff
@@ -316,45 +367,45 @@ private void sendRecord() throws InterruptedException, RemotingException, MQClie
...
-                        log.info("Successful send message to RocketMQ:{}, Topic {}", result.getMsgId(), result.getMessageQueue().getTopic());
-                        connectStatsManager.incSourceRecordWriteTotalNums();
-                        connectStatsManager.incSourceRecordWriteNums(taskConfig.getString(RuntimeConfigDefine.TASK_ID));
-                        RecordPartition partition = position.getPartition();
-                        try {
-                            if (null != partition && null != position) {
-                                Map<String, String> offsetMap = (Map<String, String>) offset.getOffset();
-                                offsetMap.put(RuntimeConfigDefine.UPDATE_TIMESTAMP, String.valueOf(sourceDataEntry.getTimestamp()));
-                                positionStorageWriter.putPosition(partition, offset);
+                try {
```
RockteMQ-AI, line 370: The async send/callback/error-handling block (~90 lines) is now duplicated between the RocketMQConverter branch and the JSON branch. Extract a shared send-with-callback helper taking the optional MessageQueue so fixes to offset-commit-on-success or failure stats don't have to be applied twice (the copy already risks divergence).

connectors/rocketmq-replicator/README.md
```diff
@@ -44,7 +44,7 @@ http://${runtime-ip}:${runtime-port}/connectors/${rocketmq-replicator-name}/stop
 注：此功能尚不成熟还需要后续版本优化
 ````
 http://${runtime-ip}:${runtime-port}/connectors/${rocketmq-replicator-name}
-?config={"connector-class":"org.apache.rocketmq.replicator.RmqMetaReplicator","source-rocketmq":"xxxx:9876","target-rocketmq":"xxxxxxx:9876","replicator-store-topic":"replicatorTopic","offset.sync.topic":"syncTopic","taskDivideStrategy":"0","white-list":"TopicTest,TopicTest2","task-parallelism":"2","source-record-converter":"org.apache.rocketmq.connect.runtime.converter.JsonConverter"}
+?config={"connector-class":"org.apache.rocketmq.replicator.RmqMetaReplicator","source-rocketmq":"xxxx:9876","target-rocketmq":"xxxxxxx:9876","target-cluster":"test1-rocketmq","source-cluster":"test1-rocketmq","replicator-store-topic":"replicatorTopic","offset.sync.topic":"syncTopic","taskDivideStrategy":"0","white-list":"TestGroup","task-parallelism":"2","source-record-converter":"org.apache.rocketmq.connect.runtime.converter.RocketMQMetaConverter"}
```
RockteMQ-AI, line 47: The meta replicator example sets source-cluster and target-cluster to the same value ("test1-rocketmq"), which is almost certainly a copy-paste mistake and confusing for users setting up cross-cluster replication; also the example still documents offset.sync.topic even though offset sync no longer flows through a topic (direct admin updates now).

RockteMQ-AI (changes requested):
### Summary
12 finding(s) to address.

### Findings
- **[CRITICAL]** `rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java:320` — Queue routing is applied unconditionally to every record in the `null == recordConverter || recordConverter instanceof RocketMQConverter` branch. ConnectRecord.getExtensions() returns null unless addExtension() was called (putExtendMsgProperty in this same method null-checks it), so `sourceDataEntry.getExtensions().getInt("queueId")` throws NPE for any record without extensions. Even when extensions exist but lack queueId/brokerName/topic, DefaultKeyValue.getInt() returns 0 and brokerName/targetTopic are null, so `new MessageQueue(null, null, 0)` is passed to producer.send() and the send fails. This is shared runtime code: every existing source connector that sets no converter or uses RocketMQConverter (not just the replicator) breaks. Guard this block (only route to a MessageQueue when topic+brokerName+queueId are all present) and fall back to producer.send(sourceMessage) otherwise.
- **[WARNING]** `rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java:270` — The RocketMQMetaConverter early-return sits inside the per-record loop and skips the `toSendRecord = null` cleanup at the end of sendRecord(). In run(), poll() is only invoked when toSendRecord is empty, so any meta-converter task that returns at least one record would busy-spin forever (100% CPU, no further polls, no position commits). Today MetaSourceTask.poll() always returns an empty list so this is dead code, but the check is converter-scoped, not record-scoped — move it before the loop (or clear the list) so the invariant can't be broken by a future task that emits records with this converter.
- **[WARNING]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/RmqSourceTask.java:182` — Adding routing metadata as plain extensions corrupts replicated messages: original message properties are copied to extensions first (line 181), then addExtension("topic"/"brokerName"/"queueId") overwrites any same-named user property, silently losing source data. Additionally, putExtendMsgProperty writes all extensions back as message properties (connect-ext-topic, connect-ext-brokerName, connect-ext-queueId), so every replicated message gains three properties the original never had. Use a dedicated prefix (e.g. connect-internal-*) or a separate mechanism for routing metadata instead of the generic "topic" key, which also risks hijacking the routing of any other connector whose records happen to carry a "topic" extension.
- **[WARNING]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java:139` — `clusterAddrTable.get(this.config.getTargetCluster())` returns null when target-cluster is not configured (older connector configs; DefaultKeyValue.put even stores null as the string "null") and `clusterNameSet.iterator()` NPEs. The broad catch swallows it with the misleading message "admin get consumer info failed" and offset sync silently does nothing — an operational trap. Validate targetCluster/targetRocketmq in validate()/start() and fail fast, and null-check the lookup with an explicit error.
- **[WARNING]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/common/Utils.java:216` — startTarMQAdminTool connects to taskConfig.getTargetRocketmq() but authenticates with SOURCE ACL credentials (isSrcAclEnable/getSrcAccessKey/getSrcSecretKey). TaskConfig has no target ACL fields even though RmqConnectorConfig does (see startTargetMQAdminTool using isTargetAclEnable), so with ACL enabled on the target cluster under different credentials every admin call fails. Also this method is a near-duplicate of startMQAdminTool/startTargetMQAdminTool — add target ACL fields to TaskConfig and reuse one implementation.
- **[WARNING]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java:151` — Offset sync matches the TARGET broker name against the SOURCE MessageQueue broker name, and WorkerSourceTask similarly sends to the same queueId/brokerName — the whole change assumes the target cluster has identical broker names and at least as many queues per topic. When names or queue counts differ, offsets are silently skipped here and data sends throw MQClientException in the runtime (records dropped). Also brokerAddrs.get(new Long(0)) returns null when no master is registered, making updateConsumeOffset fail. Validate the target route exists and log a clear warning otherwise; use Long.valueOf(0)/MixAll.MASTER_ID instead of new Long(0).
- **[WARNING]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java:102` — stop() calls tarMQAdminExt.shutdown() without a null guard. If start() throws after srcMQAdminExt started (e.g. startTarMQAdminTool failure), tarMQAdminExt is null and stop() NPEs, masking the real error, while the already-started srcMQAdminExt is never shut down (resource leak). Guard both shutdowns and release srcMQAdminExt when the target admin fails to start.
- **[WARNING]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java:115` — No tests accompany the change: the new direct updateConsumeOffset path (target-cluster lookup, broker matching, error handling), the new task-split keys in Utils.groupPartitions, and the runtime queue-routing branch are all untested, and existing tests (RmqSourceReplicatorTest, DefaultTaskDivideStrategyTest) are untouched. Please add unit tests at least for targetCluster-missing/unknown behavior and for sendRecord() with and without queueId/brokerName extensions.
- **[INFO]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java:135` — examineBrokerClusterInfo() is an expensive admin RPC invoked inside the per-group loop on every poll, and poll() is called in a tight runtime loop with no sleep, so cluster info is re-fetched G times per iteration. Hoist the call outside the groups loop and cache it (or refresh periodically); the nested group × broker × queue iteration can also be inverted so each queue is matched against a broker-name map.
- **[INFO]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java:164` — poll() now always returns an empty list — the record/schema machinery (ConnectRecord, Schema, Field, SchemaBuilder, JSONObject, FieldName, SchemaEnum imports, and the `res` variable) is dead code and poll() is purely side-effecting. Note also that OffsetSyncStore.sync() is never invoked (its consumer is never started), so convertTargetOffset degenerates to identity and updateConsumeOffset copies source offsets verbatim — wrong if target offsets diverge (e.g. replication didn't start at offset 0). Clean up the dead code and document that offset.sync.topic is no longer used by this path.
- **[INFO]** `rocketmq-connect-runtime/src/main/java/org/apache/rocketmq/connect/runtime/connectorwrapper/WorkerSourceTask.java:370` — The async send/callback/error-handling block (~90 lines) is now duplicated between the RocketMQConverter branch and the JSON branch. Extract a shared send-with-callback helper taking the optional MessageQueue so fixes to offset-commit-on-success or failure stats don't have to be applied twice (the copy already risks divergence).
- **[INFO]** `connectors/rocketmq-replicator/README.md:47` — The meta replicator example sets source-cluster and target-cluster to the same value ("test1-rocketmq"), which is almost certainly a copy-paste mistake and confusing for users setting up cross-cluster replication; also the example still documents offset.sync.topic even though offset sync no longer flows through a topic (direct admin updates now).

---

- **[WARNING]** `connectors/rocketmq-replicator/src/main/java/org/apache/rocketmq/replicator/MetaSourceTask.java:115` — No tests accompany the change: the new direct updateConsumeOffset path (target-cluster lookup, broker matching, error handling), the new task-split keys in Utils.groupPartitions, and the runtime queue-routing branch are all untested, and existing tests (RmqSourceReplicatorTest, DefaultTaskDivideStrategyTest) are untouched. Please add unit tests at least for targetCluster-missing/unknown behavior and for sendRecord() with and without queueId/brokerName extensions. _(line outside diff)_

RockteMQ-AI (approved):
### Summary
Fixes message synchronization exceptions in rocketmq-replicator. Improves error handling and retry logic.

LGTM.

---

Issue #189, opened 2022-07-12
java.lang.AbstractMethodError: io.openmessaging.connector.api.component.connector.Connector.start


JasonRoyGoGo:
2022-07-12 10:38:58 ERROR qtp1848289347-34 - Exception occurred while servicing http-request
java.lang.AbstractMethodError: io.openmessaging.connector.api.component.connector.Connector.start(Lio/openmessaging/KeyValue;)V
	at org.apache.rocketmq.connect.runtime.service.memory.MemoryConfigManagementServiceImpl.putConnectorConfig(MemoryConfigManagementServiceImpl.java:158)
	at org.apache.rocketmq.connect.runtime.rest.RestHandler.handleCreateConnector(RestHandler.java:145)
	at io.javalin.security.SecurityUtil.noopAccessManager(SecurityUtil.kt:22)
	at io.javalin.Javalin.lambda$addHandler$0(Javalin.java:606)
	at io.javalin.core.JavalinServlet$service$2$1.invoke(JavalinServlet.kt:46)
	at io.javalin.core.JavalinServlet$service$2$1.invoke(JavalinServlet.kt:17)
	at io.javalin.core.JavalinServlet$service$1.invoke(JavalinServlet.kt:143)
	at io.javalin.core.JavalinServlet$service$2.invoke(JavalinServlet.kt:41)
	at io.javalin.core.JavalinServlet.service(JavalinServlet.kt:107)
	at io.javalin.core.util.JettyServerUtil$initialize$httpHandler$1.doHandle(JettyServerUtil.kt:72)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:203)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:480)
	at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1668)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:201)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1247)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:144)
	at org.eclipse.jetty.server.handler.HandlerList.handle(HandlerList.java:61)
	at org.eclipse.jetty.server.handler.StatisticsHandler.handle(StatisticsHandler.java:174)
[...]

PR #194, opened 2022-07-14
[ISSUE #189]Upgrade rocketmq-replicator connector API to v0.1.3

sunxiaojian:
Upgrade rocketmq-replicator connector API to v0.1.3


odbozhou (approved):

RockteMQ-AI (approved):
### Summary

Upgrades the replicator connector to use the `openmessaging-connector 0.1.3` API. The main change is consolidating the lifecycle: `init()` + `start(ConnectorContext)` → `start(KeyValue)`. Also removes empty `pause()`/`resume()` overrides since the default no-op implementation is sufficient.

### Findings
- **[Info]** `RocketMQSourceConnector.java:48` — The `start(KeyValue)` method now takes the config directly, which is cleaner than the previous two-phase init. The `ConnectorClassSetter.setConnectorClass(config)` call is correctly placed before `super.start(config)`.
- **[Info]** Removing the empty `pause()`/`resume()` overrides is correct — the base class provides no-op defaults.

LGTM.

PR #231, opened 2022-08-03
fix bug in runconnect.sh

ingdex:
to fix 2 bug in runconnect.sh
1.check_java_version cannot parse java version to a right number.
2.```sed``` command can not work well in maxOS, so I replease it with ```tr``` instead.

I followed quick start in README.md, however, when I run ```sh bin/connect-standalone.sh -c conf/connect-standalone.conf & ```
I got error as follow:
```
(standard_in) 1: parse error
sed: 1: ":a;N;s/\n/:/;ba;": unused label 'a;N;s/\n/:/;ba;'
(standard_in) 1: parse error
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option PermSize=128m; support was removed in 8.0
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=320m; support was removed in 8.0
Unrecognized option: -Xlog:gc:/dev/shm/mq_gc_%p.log
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```
The two parse error is caused by check_java_version function in shell script.
And another problem is than ```sed``` command in shell script will cause "unused label" error in macOS, so I used ```tr``` instead to get the correct CLASSPATH to fixed it by the way.

### DIFF
--- a/distribution/bin/runconnect.sh
+++ b/distribution/bin/runconnect.sh
@@ -32,7 +32,7 @@ check_java_version ()
 {
   _java=$1
   _version=$2
-  version=$("$_java" -version 2>&1 | awk -F '"' '/version/ {print $2}' | grep -o '^[0-9.]\+')
+  version=$("$_java" -version 2>&1 | awk -F '"' '/version/ {print $2}' | grep -o '^[0-9]\+\.\+[0-9]\+')
   flag="true"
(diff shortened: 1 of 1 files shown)


Oliverwqcwrw (approved):
Well, I tried the previous script and it does have this problem on the MAC. Thank you for your contribution. The community really needs developers like you

PR #516, opened 2023-06-18
improve jdbc connector docs

Ziy1-Tan:

- The package name has changed, the `connector.class` needs to be updated

## Brief changelog

- update the jdbc sink class name


joeCarf (approved):
LGTM

PR #533, opened 2024-05-17
Feature:支持CloudEvent事件写入OSS,OSS存储格式时间分区功能,按时间窗口和事件内容信息大小窗口攒批发送功能

liiuzq-xiaobai:

支持CloudEvent事件写入OSS,OSS存储格式时间分区功能,按时间窗口和事件内容信息大小窗口攒批发送功能

### DIFF
--- a/connectors/aliyun/rocket-connect-oss/pom.xml
+++ b/connectors/aliyun/rocket-connect-oss/pom.xml
@@ -1,0 +1,13 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.rocketmq</groupId>
+        <artifactId>rocketmq-connect</artifactId>
+        <version>0.0.1-SNAPSHOT</version>
+    </parent>
+    <groupId>org.example</groupId>
+    <artifactId>rocket-connect-oss</artifactId>
+    <name>Archetype - rocket-connect-oss</name>
+    <url>http://maven.apache.org</url>
+</project>
--- a/connectors/aliyun/rocket-connect-oss/pom.xml
+++ b/connectors/aliyun/rocket-connect-oss/pom.xml
@@ -1,13 +1,246 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Licensed to the Apache Software Foundation (ASF) under one or more contributor
+	license agreements. See the NOTICE file distributed with this work for additional
+	information regarding copyright ownership. The ASF licenses this file to
+	You under the Apache License, Version 2.0 (the "License"); you may not use
+	this file except in compliance with the License. You may obtain a copy of
+	the License at http://www.apache.org/licenses/LICENSE-2.0 Unless required
+	by applicable law or agreed to in writing, software distributed under the
+	License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS
+	OF ANY KIND, either express or implied. See the License for the specific
+	language governing permissions and limitations under the License. -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-    <parent>
-        <groupId>org.apache.rocketmq</groupId>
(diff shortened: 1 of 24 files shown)


RockteMQ-AI (commented):
### Summary
This PR adds an OSS (Object Storage Service) sink connector for RocketMQ Connect, supporting CloudEvent writes to Alibaba Cloud OSS with time-partitioned storage and configurable batch strategies (default, time-window, size-window). The overall architecture is sound, but there are several critical correctness issues that need to be addressed before merging.

### Findings
- **[Critical]** `CloudEventManager.java:598` — Validation checks `ACCESS_KEY_SECRET` twice instead of checking `ACCESS_KEY_ID` first; blank access key ID passes validation
- **[Critical]** `SizeWindowEventProcessor.java:752` — `batchSize` and `MAX_BATCH_SIZE` are static mutable fields shared across all instances; should be instance fields
- **[Critical]** `SizeWindowEventProcessor.java:782` — Race condition: non-atomic check-then-send-then-reset in `send()` allows concurrent threads to trigger duplicate sends or lose records
- **[Warning]** `SizeWindowEventProcessor.java:794` — `asyncSendBatch()` is actually synchronous; misleading name
- **[Warning]** `TimeWindowEventProcessor.java:859` — Static `ScheduledExecutorService` shared across instances; `shutdown()` doesn't flush pending records
- **[Warning]** `SizeWindowEventProcessor.java:780` — Buffered records are lost on `shutdown()`; violates at-least-once delivery guarantee
- **[Info]** No integration tests for the connector; only a basic `TimeWindowEventProcessorTest` exists
- **[Info]** `OSSSinkTask.validate()` calls `cloudEventManager.validate()` but `cloudEventManager` could be null if called before initialization
- **[Info]** `connectRecord.getData().toString()` assumes meaningful `toString()` — may produce garbage for complex data types

### Suggestions
1. Fix the validation bug (copy-paste error on `ACCESS_KEY_SECRET`)
2. Make `batchSize`, `MAX_BATCH_SIZE`, and `scheduler` instance fields instead of static
3. Add proper synchronization around the batch threshold check and send
4. Implement graceful shutdown that flushes pending records before closing
5. Rename `asyncSendBatch()` to `sendBatch()` or make it truly async
6. Add integration tests that verify end-to-end data flow and at-least-once semantics
7. Consider using `Preconditions.checkNotNull()` for cleaner null checks

### Cross-repo Note
This PR is part of the `apache/rocketmq-connect` ecosystem. The connector follows the OpenMessaging Connect API pattern. Ensure compatibility with the Connect runtime's task lifecycle (start/stop/pause/resume).

---

- **[CRITICAL]** `connectors/aliyun/rocket-connect-oss/src/main/java/org/apache/rocketmq/connect/oss/service/CloudEventManager.java:598` — **[Critical]** Validation bug: `ACCESS_KEY_SECRET` is checked twice (lines 598-599), but `ACCESS_KEY_ID` is never validated. This means a blank access key ID would pass validation and cause runtime failures.

Should be:
```java
if (StringUtils.isBlank(config.getString(TaskConfig.ACCESS_KEY_ID))
        || StringUtils.isBlank(config.getString(TaskConfig.ACCESS_KEY_SECRET))
```

This is a copy-paste bug that will cause authentication failures at runtime. _(line outside diff)_
- **[CRITICAL]** `connectors/aliyun/rocket-connect-oss/src/main/java/org/apache/rocketmq/connect/oss/service/SizeWindowEventProcessor.java:752` — **[Critical]** Static mutable state: `batchSize` and `MAX_BATCH_SIZE` are declared `static`, meaning they're shared across all instances of this class. If multiple sink tasks run in the same JVM (common in Connect frameworks), they'll interfere with each other's batch counting.

These should be instance fields, not static. Same issue exists in `TimeWindowEventProcessor` with the static `scheduler`. _(line outside diff)_
- **[CRITICAL]** `connectors/aliyun/rocket-connect-oss/src/main/java/org/apache/rocketmq/connect/oss/service/SizeWindowEventProcessor.java:782` — **[Critical]** Race condition: The check `batchSize.get() + recordLength > MAX_BATCH_SIZE` followed by `asyncSendBatch()` and `batchSize.set(0)` is not atomic. Multiple threads calling `send()` concurrently could:
1. Both see the threshold exceeded
2. Both call `asyncSendBatch()`
3. Both reset the counter

This leads to duplicate sends, lost records, or inconsistent state. Use synchronization or `AtomicLong.accumulateAndGet()` with proper locking. _(line outside diff)_
- **[WARNING]** `connectors/aliyun/rocket-connect-oss/src/main/java/org/apache/rocketmq/connect/oss/service/SizeWindowEventProcessor.java:794` — **[Warning]** Misleading method name: `asyncSendBatch()` is actually synchronous (calls `ossClientProvider.getOssClient().putObject(...)` directly). The "async" prefix suggests non-blocking behavior that doesn't exist. Either rename to `sendBatch()` or make it truly async with a thread pool. _(line outside diff)_
- **[WARNING]** `connectors/aliyun/rocket-connect-oss/src/main/java/org/apache/rocketmq/connect/oss/service/TimeWindowEventProcessor.java:859` — **[Warning]** Static `ScheduledExecutorService`: The `scheduler` is static, so all instances share the same thread pool. This is problematic if multiple tasks run concurrently. Also, `shutdown()` calls `scheduler.shutdown()` but doesn't await termination or flush pending records, risking data loss on task restart. _(line outside diff)_
- **[WARNING]** `connectors/aliyun/rocket-connect-oss/src/main/java/org/apache/rocketmq/connect/oss/service/SizeWindowEventProcessor.java:780` — **[Warning]** Data loss risk: When `shutdown()` is called, any records buffered in the queue are not flushed to OSS. This violates the at-least-once delivery guarantee expected from a sink connector. The shutdown sequence should:
1. Stop accepting new records
2. Flush the current batch
3. Wait for in-flight uploads to complete _(line outside diff)_

